PR #2, opened 2018-09-20
Covariance

LukeMathWalker:
A basic implementation for the computation of the covariance matrix starting from a 2-dimensional array of observations.

Ideally, this should be implemented for any n-dimensional array with n>=2, treating two user-specified axes as `random_variable` axis and `observation` axis, while using the remaining ones as batch axes. To implement the generalized version we have to be able to do an efficient [batched matrix multiplication](https://devblogs.nvidia.com/cublas-strided-batched-matrix-multiply/) - the implementation of this routine is currently missing in ndarray.


LukeMathWalker:
The compiler complained about the lifetime of type `A` and to "calm it" I slapped `'static` in front of `A`. Can you help understand what the issue was? @jturner314

src/correlation.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// where
+    ///
+    /// ```text
+    ///     1   n
+    /// x̅ = ―   ∑ xᵢ
+    ///     n  i=1
+    /// ```
+    /// and similarly for ̅y. 
+    ///
+    /// **Panics** if `ddof` is greater than or equal to the number of observations.
```
jturner314: It's worth noting that the implementation also panics if converting the number of observations to `A` with `FromPrimitive` fails.

PR #20, opened 2019-01-05
Means

LukeMathWalker:
A basic implementation of arithmetic, harmonic and geometric mean.

Even though `ndarray` exposes `mean_axis` it does not provide `mean`, hence I have added it to the PR. Should I contribute it back to `ndarray` @jturner314?

### DIFF
--- a/src/summary_statistics/means.rs
+++ b/src/summary_statistics/means.rs
@@ -82,20 +82,46 @@ mod tests {
     use ndarray::Array1;
 
     #[test]
-    fn test_mean_with_nan_values() {
+    fn test_means_with_nan_values() {
         let a = array![f64::NAN, 1.];
         assert!(a.mean().unwrap().is_nan());
+        assert!(a.harmonic_mean().unwrap().is_nan());
     }
 
     #[test]
-    fn test_mean_with_empty_array_of_floats() {
+    fn test_means_with_empty_array_of_floats() {
         let a: Array1<f64> = array![];
         assert!(a.mean().is_err());
+        assert!(a.harmonic_mean().is_err());
     }
 
     #[test]
-    fn test_mean_with_empty_array_of_noisy_floats() {
+    fn test_means_with_empty_array_of_noisy_floats() {
         let a: Array1<N64> = array![];
         assert!(a.mean().is_err());
+        assert!(a.harmonic_mean().is_err());
+    }
+
+    #[test]
+    fn test_means_with_array_of_floats() {
+        let a: Array1<f64> = array![
+            0.99889651, 0.0150731 , 0.28492482, 0.83819218, 0.48413156,
+            0.80710412, 0.41762936, 0.22879429, 0.43997224, 0.23831807,
+            0.02416466, 0.6269962 , 0.47420614, 0.56275487, 0.78995021,
+            0.16060581, 0.64635041, 0.34876609, 0.78543249, 0.19938356,
+            0.34429457, 0.88072369, 0.17638164, 0.60819363, 0.250392  ,
+            0.69912532, 0.78855523, 0.79140914, 0.85084218, 0.31839879,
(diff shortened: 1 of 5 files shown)


src/summary_statistics/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// [`harmonic mean`]: https://en.wikipedia.org/wiki/Harmonic_mean
+    fn harmonic_mean(&self) -> Result<A, &'static str>
+        where
+            A: Float + FromPrimitive + PartialEq;
+
+    /// Returns the [`geometric mean`] `GM(X)` of all elements in the array:
+    ///
+    /// ```text
+    ///           n
```
jturner314: I like big parentheses better in this case:

```rust
    ///         ⎛ n   ⎞¹⁄ₙ
    /// GM(X) = ⎜ ∏ xᵢ⎟
    ///         ⎝i=1  ⎠
```

src/summary_statistics/means.rs
```diff
@@ -0,0 +1,106 @@
...
+
+
+impl<A, S, D> SummaryStatisticsExt<A, S, D> for ArrayBase<S, D>
+where
+    S: Data<Elem = A>,
+    D: Dimension,
+{
+    fn mean(&self) -> Result<A, &'static str>
+    where
+        A: Clone + FromPrimitive + Add<Output=A> + Div<Output=A> + PartialEq + Zero
```
jturner314: What it the `PartialEq` bound used for?

src/summary_statistics/means.rs
```diff
@@ -0,0 +1,106 @@
...
+                .expect("Converting number of elements to `A` must not fail.");
+            Ok(self.sum() / n_elements)
+        }
+    }
+
+    fn harmonic_mean(&self) -> Result<A, &'static str>
+    where
+        A: Float + FromPrimitive + PartialEq,
+    {
+        if self.len() == 0 {
```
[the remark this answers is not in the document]
LukeMathWalker: It was implemented separately to return a more precise error message, but if we use `Option` I can get rid of it.

src/summary_statistics/means.rs
LukeMathWalker: Probably it survived some early iterations I was doing. I got rid of it.

LukeMathWalker:
> I do think we should add .mean() to ndarray since ndarray already has .sum() and .mean_axis().

I agree - but the semantic of `mean_axis` is different at the moment: it panics instead of returning a `None` if the array is empty. That might have to be changed.

> It's unfortunate that harmonic_mean and geometric_mean allocate a temporary array for the result of the map. However, for the time being, I don't see a good way to avoid this while still taking advantage of the pairwise implementation of .sum() (rust-ndarray/ndarray#577).

It's exactly what I thought (and mean calculation is why I drafted the pairwise summation PR in the first place).

LukeMathWalker:
CI on Rust 1.30 is failing with `error: Edition 2018 is unstable and only available for nightly builds of rustc.` for `noisy_floats`. Curious.

jturner314:
Everything looks good, so I merged the PR. If `.mean()` gets added to `ndarray`, we can remove it from `ndarray-stats`.

> I agree - but the semantic of `mean_axis` is different at the moment: it panics instead of returning a `None` if the array is empty. That might have to be changed.

Will you please submit a PR to `ndarray` to add `.mean()`? I lean towards changing `mean_axis` to return `None` instead of making `mean` divide by zero, because of [this reasoning](https://github.com/rust-ndarray/ndarray/issues/512#issuecomment-433239806). However, for floating-point values, division by zero would be more convenient since it would just result in NaN instead of panicking. What do you think? I'll be interested to hear @bluss's opinion too.

bluss:
@jturner314 Your rule makes sense, sounds good to me to return an Option. I think it's quite frequent anyway that you want to compute a mean and if empty, get zero, and the option makes it simple to do that.

PR #23, opened 2019-01-15
Central moments

LukeMathWalker:
Computation of central moments of arbitrary order.

Would it be worth to add another method that takes the mean as a parameter to avoid re-computing it if the user already has its value @jturner314?

### DIFF
--- a/src/summary_statistics/means.rs
+++ b/src/summary_statistics/means.rs
@@ -2,7 +2,7 @@ use super::SummaryStatisticsExt;
 use crate::errors::EmptyInput;
 use ndarray::{ArrayBase, Data, Dimension};
 use num_integer::IterBinomial;
-use num_traits::{Float, FromPrimitive, ToPrimitive, Zero};
+use num_traits::{Float, FromPrimitive, Zero};
 use std::ops::{Add, Div};
 
 impl<A, S, D> SummaryStatisticsExt<A, S, D> for ArrayBase<S, D>
@@ -127,9 +127,7 @@ where
 {
     let n_elements =
         A::from_usize(a.len()).expect("Converting number of elements to `A` must not fail");
-    let order = order
-        .to_i32()
-        .expect("Moment order must not overflow `i32`.");
+    let order = order as i32;
 
     // When k=0, we are raising each element to the 0th power
     // No need to waste CPU cycles going through the array
@@ -317,8 +315,7 @@ mod tests {
         let order = 10;
         let central_moments = a.central_moments(order).unwrap();
         for i in 0..=order {
-            assert_eq!(a.central_moment(i).unwrap(), central_moments[i as usiz
-            e]);
+            assert_eq!(a.central_moment(i).unwrap(), central_moments[i as usize]);
         }
     }
 
--- a/src/summary_statistics/means.rs
+++ b/src/summary_statistics/means.rs
@@ -321,7 +321,8 @@ mod tests {
 
     #[test]
     fn test_kurtosis_and_skewness_is_none_with_empty_array_of_floats() {
-        let a: Array1<f64> = array![];
+        let a: Array1<f64> = array![]
+            ;
         assert_eq!(a.skewness(), Err(EmptyInput));
         assert_eq!(a.kurtosis(), Err(EmptyInput));
     }
--- a/src/summary_statistics/means.rs
+++ b/src/summary_statistics/means.rs
@@ -42,16 +42,48 @@ where
         A: Float + FromPrimitive
     {
         let mean = self.mean();
-        if mean.is_none() {
-            None
-        } else {
-            match n {
-                0 => Some(A::one()),
-                1 => Some(A::zero()),
-                n => unimplemented!()
+        match mean {
+            None => None,
+            Some(mean) => {
+                match n {
+                    0 => Some(A::one()),
+                    1 => Some(A::zero()),
+                    n => {
+                        let n_elements = A::from_usize(self.len()).
+                            expect("Converting number of elements to `A` must not fail");
+                        let shifted_array = self.map(|x| x.clone() - mean);
+                        let correction_term = -shifted_array.sum() / n_elements;
+                        let coefficients: Vec<A> = (0..n).map(
+                            |k| A::from_usize(binomial_coefficient(n, k)).unwrap() *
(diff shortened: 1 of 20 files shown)


LukeMathWalker:
Once pairwise summation gets merged in ndarray, this will actually be a corrected pairwise two-pass algorithm for central moments.

LukeMathWalker:
Added kurtosis and skewness computation.

src/summary_statistics/means.rs
```diff
@@ -88,16 +256,114 @@ mod tests {
         // Computed using SciPy
         let expected_geometric_mean = 0.4345897639796527;
 
-        abs_diff_eq!(a.mean().unwrap(), expected_mean, epsilon = f64::EPSILON);
-        abs_diff_eq!(
+        assert_abs_diff_eq!(a.mean().unwrap(), expected_mean, epsilon = 1e-6);
```
jturner314: This `epsilon` is fairly large. Is it due to `ndarray` not having pairwise summation yet?

src/summary_statistics/mod.rs
```diff
@@ -60,6 +60,83 @@ where
...
+    /// If the array is empty, `None` is returned.
+    ///
+    /// The *p*-th central moment is computed using a corrected two-pass algorithm (see Section 3.5
+    /// in [Pébay et al., 2016]). Complexity is *O(np)* when *n >> p*, *p > 1*.
+    ///
+    /// **Panics** if `A::from_usize()` fails to convert the number of elements in the array.
+    ///
+    /// [central moment]: https://en.wikipedia.org/wiki/Central_moment
+    /// [Pébay et al., 2016]: https://www.osti.gov/pages/servlets/purl/1427275
+    fn central_moment(&self, order: usize) -> Option<A>
```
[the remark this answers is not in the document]
LukeMathWalker: I used `usize` out of convenience, but given that we are performing a conversion to `i32` it's better to go for a `u32`, you're right.

src/summary_statistics/means.rs
LukeMathWalker: I have tuned all epsilons to be as close as possible to `1e-12`, stopping when the assertion fails.
It seems our mean estimate agrees with NumPy's with a delta of `1e-9` - I'd say that the lack of pairwise summation is a big suspect. It would be nice to have something like [decimal](https://docs.python.org/3/library/decimal.html#module-decimal) in Rust to get a sense of the real error.

jturner314: > I have tuned all epsilons to be as close as possible to `1e-12`, stopping when the assertion fails.

Okay, thanks. That's helpful to get a better understanding of the precision.

> It would be nice to have something like [decimal](https://docs.python.org/3/library/decimal.html#module-decimal) in Rust to get a sense of the real error.

For this particular case, `decimal` gives the following for the standard and harmonic mean:

```python
>>> from decimal import *
>>> getcontext().prec = 100
>>> data = [
...     Decimal('0.99889651'), Decimal('0.0150731'), Decimal('0.28492482'), Decimal('0.83819218'), Decimal('0.48413156'), Decimal('0.80710412'), Decimal('0.41762936'),
...     Decimal('0.22879429'), Decimal('0.43997224'), Decimal('0.23831807'), Decimal('0.02416466'), Decimal('0.6269962'), Decimal('0.47420614'), Decimal('0.56275487'),
...     Decimal('0.78995021'), Decimal('0.16060581'), Decimal('0.64635041'), Decimal('0.34876609'), Decimal('0.78543249'), Decimal('0.19938356'), Decimal('0.34429457'),
...     Decimal('0.88072369'), Decimal('0.17638164'), Decimal('0.60819363'), Decimal('0.250392'), Decimal('0.69912532'), Decimal('0.78855523'), Decimal('0.79140914'),
...     Decimal('0.85084218'), Decimal('0.31839879'), Decimal('0.63381769'), Decimal('0.22421048'), Decimal('0.70760302'), Decimal('0.99216018'), Decimal('0.80199153'),
...     Decimal('0.19239188'), Decimal('0.61356023'), Decimal('0.31505352'), Decimal('0.06120481'), Decimal('0.66417377'), Decimal('0.63608897'), Decimal('0.84959691'),
...     Decimal('0.43599069'), Decimal('0.77867775'), Decimal('0.88267754'), Decimal('0.83003623'), Decimal('0.67016118'), Decimal('0.67547638'), Decimal('0.65220036'),
...     Decimal('0.68043427')
... ]
>>> sum(data) / len(data)
Decimal('0.5475494054')
>>> 1 / (sum(1/x for x in data) / len(data))
Decimal('0.2179009364951495638060724591285330338229747648318613758290505954023583698639515519076072129091188061')
```

In Rust, we could add a dev-dependency on [`rug`](https://crates.io/crates/rug) to compute the correct result to the necessary precision, but it's not necessary for this PR.

LukeMathWalker: That's interesting - using the results from

LukeMathWalker: That's reassuring: using the results from `decimal` it turns out our computation is correct up to `f64::EPSILON` in this specific case.

LukeMathWalker: We should definitely look into using something like `rug` to keep in check our numerical error. On the side, comparing with SciPy/NumPy reassures me that the algorithm implementation is roughly correct, even if due to the respective numerical error we do not agree on the smaller digits.

LukeMathWalker:
I have changed the return type to match what we decided in #25 - it should be ready to be merged now @jturner314

LukeMathWalker:
I agree - I changed it to `u16` @jturner314.

src/summary_statistics/means.rs
```diff
@@ -36,15 +37,163 @@ where
...
+/// **Panics** if `A::from_usize()` fails to convert the number of elements in the array.
+fn moments<A, S, D>(a: ArrayBase<S, D>, order: u16) -> Vec<A>
+where
+    A: Float + FromPrimitive,
+    S: Data<Elem = A>,
+    D: Dimension,
+{
+    let n_elements =
+        A::from_usize(a.len()).expect("Converting number of elements to `A` must not fail");
+    let order = order
```
jturner314: Now that `order` is `u16`, we can remove this.

LukeMathWalker: True, it can't fail anymore. I'll merge once the CI has finished :+1:

PR #24, opened 2019-01-27
Entropy

LukeMathWalker:
This PR adds a new extension trait for Information Theory quantities - so far, entropy, cross entropy and Kullback-Leibler divergence.

Open question: returning an `Option` for cross entropy and KL divergence conflates two different "failure" modes - `None` when the arrays are empty, `None` when we have a dimension mismatch.
Would it be better to return a `Result<Option<A>, E>`, with a custom `DimensionMismatch` error type?


src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// The Kullback-Leibler divergence is defined as:
+    ///
+    /// ```text
+    ///              n
+    /// Dₖₗ(p,q) = - ∑ pᵢ ln(qᵢ/pᵢ)
+    ///             i=1
+    /// ```
+    ///
+    /// If the arrays are empty or their lengths are not equal, `None` is returned.
```
jturner314: I agree that it's not good to combine these two cases. A few options:

* `Result<Option<A>, DimensionMismatch>` would be fine.
* I think `Result<A, ShapeError>` is a bit cleaner (where `ShapeError` is an enum with `Empty` and `ShapeMismatch` variants).
* Another approach is `Option<A>`, returning `None` in the empty case and panicking if `q` cannot be broadcast to the shape of `p`. I generally avoid panicking, but I think this is the best option in this case because:

  * It's more consistent with `ndarray`'s methods that operate on two arrays.
  * If the caller passes in arrays of mismatched shapes, it's clearly a bug in their code.
  * If the arrays have mismatched shapes, it's hard to see how the caller would recover from this error case.

  The question then is why we should return `None` in the empty case instead of panicking. I think this makes sense because it's much more likely than a shape mismatch, doesn't really indicate a bug in the caller, and is more likely to be recoverable.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// The Kullback-Leibler divergence is a measure used in [Information Theory]
+    /// to describe the relationship between two probability distribution: it only make sense
+    /// when each array sums to 1 with entries between 0 and 1 (extremes included).
+    ///
+    /// By definition, *pᵢ ln(qᵢ/pᵢ)* is set to 0 if *pᵢ* is 0.
+    ///
+    /// [Kullback-Leibler divergence]: https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence
+    /// [Information Theory]: https://en.wikipedia.org/wiki/Information_theory
+    fn kl_divergence(&self, q: &Self) -> Option<A>
```
jturner314: `q: &Self` is unnecessarily restrictive. The signature should be

```rust
fn kl_divergence<S2>(&self, q: &ArrayBase<S2, D>) -> Option<A>
where
    S2: Data<Elem = A>,
```

or if we allow broadcasting,

```rust
fn kl_divergence<S2, E>(&self, q: &ArrayBase<S2, E>) -> Option<A>
where
    S2: Data<Elem = A>,
    E: Dimension,
```

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// The cross entropy is defined as:
+    ///
+    /// ```text
+    ///            n
+    /// H(p,q) = - ∑ pᵢ ln(qᵢ)
+    ///           i=1
+    /// ```
+    ///
+    /// If the arrays are empty or their lengths are not equal, `None` is returned.
```
jturner314: See the comment above about shape mismatches in `kl_divergence`.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// The cross entropy is often used as an objective/loss function in
+    /// [optimization problems], including [machine learning].
+    ///
+    /// By definition, *pᵢ ln(qᵢ)* is set to 0 if *pᵢ* is 0.
+    ///
+    /// [cross entropy]: https://en.wikipedia.org/wiki/Cross-entropy
+    /// [Information Theory]: https://en.wikipedia.org/wiki/Information_theory
+    /// [optimization problems]: https://en.wikipedia.org/wiki/Cross-entropy_method
+    /// [machine learning]: https://en.wikipedia.org/wiki/Cross_entropy#Cross-entropy_error_function_and_logistic_regression
+    fn cross_entropy(&self, q: &Self) -> Option<A>
```
jturner314: See the comment above about the type of `q` for `kl_divergence`.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// ```text
+    ///       n
+    /// S = - ∑ xᵢ ln(xᵢ)
+    ///      i=1
+    /// ```
+    ///
+    /// If the array is empty, `None` is returned.
+    ///
+    /// **Panics** if any element in the array is negative.
```
jturner314: SciPy [defines the entropy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.entr.html#scipy.special.entr) for a negative element as negative infinity instead of panicking. This is nice because it avoids panics, but I'm not sure whether defining it as negative infinity really makes sense, though. Thoughts?

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// ```text
+    ///              n
+    /// Dₖₗ(p,q) = - ∑ pᵢ ln(qᵢ/pᵢ)
+    ///             i=1
+    /// ```
+    ///
+    /// If the arrays are empty or their lengths are not equal, `None` is returned.
+    ///
+    /// **Panics** if any element in *q* is negative and taking the logarithm of a negative number
```
jturner314: SciPy [defines the relative entropy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.rel_entr.html) for a negative element as infinity instead of panicking. This is nice because it avoids panics, but I'm not sure whether defining it as infinity really makes sense, though. Thoughts?

On another note, the docs should indicate that this happens when an element of `p` is negative in addition to an element of `q` being negative.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+            ).sum();
+            Some(-entropy)
+        }
+    }
+
+    fn kl_divergence(&self, q: &Self) -> Option<A>
+        where
+            A: Float
+    {
+        if (self.len() == 0) | (self.len() != q.len()) {
```
jturner314: The shapes should match, not just the number of elements.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+            ).collect::<Array1<A>>().sum();
+            Some(-kl_divergence)
+        }
+    }
+
+    fn cross_entropy(&self, q: &Self) -> Option<A>
+        where
+            A: Float
+    {
+        if (self.len() == 0) | (self.len() != q.len()) {
```
jturner314: The shapes should match, not just the number of elements.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    fn kl_divergence(&self, q: &Self) -> Option<A>
+        where
+            A: Float
+    {
+        if (self.len() == 0) | (self.len() != q.len()) {
+            None
+        } else {
+            let kl_divergence: A = self.iter().zip(q.iter()).map(
```
jturner314: I'd prefer to use `ndarray::Zip` because it should have better performance.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+                |(p, q)| {
```
jturner314: Just a style thing -- when I have a closure that takes references to a type that implements `Copy`, I usually write the closure like this:

```rust
|(&p, &q)| {
    if p == A::zero() {
        p
    } else {
        p * (q / p).ln()
    }
}
```

so that I don't have to explicitly dereference the parameters in the body every time I use them.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+                    if *p == A::zero() {
+                        A::zero()
+                    } else {
+                        *p * (*q / *p).ln()
+                    }
+                }
+            ).collect::<Array1<A>>().sum();
```
jturner314: SciPy [sums only along axis 0](https://github.com/scipy/scipy/blob/v1.2.1/scipy/stats/_distn_infrastructure.py#L2515-L2556). To be honest, I'm not really sure why they decided to do it that way, but I thought I'd point it out.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// If the arrays are empty or their lengths are not equal, `None` is returned.
+    ///
+    /// **Panics** if any element in *q* is negative and taking the logarithm of a negative number
+    /// is a panic cause for `A`.
+    ///
+    /// ## Remarks
+    ///
+    /// The cross entropy is a measure used in [Information Theory]
+    /// to describe the relationship between two probability distribution: it only make sense
```
jturner314:
```suggestion
    /// to describe the relationship between two probability distributions: it only makes sense
```

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    fn cross_entropy(&self, q: &Self) -> Option<A>
+        where
+            A: Float
+    {
+        if (self.len() == 0) | (self.len() != q.len()) {
+            None
+        } else {
+            let cross_entropy: A = self.iter().zip(q.iter()).map(
```
jturner314: See comments above about `ndarray::Zip` and closure formatting style.

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+                |(p, q)| {
+                    if *p == A::zero() {
+                        A::zero()
+                    } else {
+                        *p * q.ln()
+                    }
+                }
+            ).collect::<Array1<A>>().sum();
```
jturner314: See comment above about axis/axes to sum over.

src/entropy.rs
LukeMathWalker: Mathematically, it does not and it can turn out into a silent failure. I'd rather be vocal and upfront about invalid values (return an error/panic).

src/entropy.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// If the array is empty, `None` is returned.
+    ///
+    /// **Panics** if any element in the array is negative.
+    ///
+    /// ## Remarks
+    ///
+    /// The entropy is a measure used in [Information Theory]
+    /// to describe a probability distribution: it only make sense
+    /// when the array values sum to 1, with each entry between
```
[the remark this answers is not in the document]
LukeMathWalker: I thought about SciPy's behaviour, but normalization can introduce numerical errors before even starting the entropy computation (especially with values of small absolute magnitude). 
To proceed properly one would have to first sum the array, check if it sums to 1, then normalize if required, then compute the entropy. It's possible, but I'd avoid forcing everyone to pay the penalty - I'd rather implement a `checked` variant of the same method with this behaviour if it's of interest.

src/entropy.rs
jturner314: Okay, sounds good to me.

src/entropy.rs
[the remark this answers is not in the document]
jturner314: Okay, that makes sense. I hadn't considered that normalization could introduce numerical errors.

src/entropy.rs
LukeMathWalker: I'd rather go for `Result<A, ShapeError>` - in the end, the caller can choose how to recover from errors (free to call `unwrap` or `expect`).
I can foresee various scenarios for both failure modes:
- empty because you filtered some values out of another array/vec, but now nothing is left;
- shape mismatch, because you read two different samples from files on disk which you expected to have the same number of values but... they don't (real life stories here 😅).

LukeMathWalker: To provide a plausible use case: it might be impossible to recover (let the program resume its expected execution flow) but there might be some actions one might want to perform before panicking for example (logging is the first that comes to my mind or sending a request somewhere in a web application context).
One can always catch the panic using stuff like [this](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) but it's less pleasant and not clear at a first glance at the function signature.

jturner314: Okay, that makes sense. We have to return an `Option`/`Result` anyway, so we might as well handle all the error cases without panicking. Along the same lines, should we return an error on negative values instead of panicking? I'd expect negative values to be a bigger issue than shape mismatches, and we're already checking if values are zero.

LukeMathWalker: On that one I am torn, because the cost of checking for negative values scales with the number of elements given that it's an additional check. Do you think it's worth it?

src/entropy.rs
LukeMathWalker: I have always some problems when using `Zip` - can you help understand what the compiler is complaining about?
image

(`.fold_while` is more appropriate, but I was doing a first step just to get `Zip` there)

LukeMathWalker: I got it to work by changing
`let mut temp = ArrayBase::zeros(self.shape());`
into
`let mut temp = Array::zeros(self.raw_dim());`
but I really want to understand the reason behind the extremely different outcome :sweat_smile:

src/entropy.rs
LukeMathWalker: It might have something to do with a convention on which axis represents the number of samples?
Overall I find SciPy's support for information theory quantities to be quite poor from an API perspective, I have mostly drawn inspiration from [StatsBase.jl](http://juliastats.github.io/StatsBase.jl/stable/scalarstats.html#Entropy-and-Related-Functions-1).

src/entropy.rs
LukeMathWalker: Thinking over it again, `Result<Option<A>, DimensionMismatch>` is probably the best signature, given that most of our methods return an `Option<A>` with `None` when arrays are empty. Consistency is probably worth the double wrap (unless we want to use `DimensionMismatch::Empty` instead of `Option` in the rest of the API).

jturner314: I didn't realize earlier that the `ln` of a negative number is `NaN` (except for the noisy float types, which panic). That behavior is fine with me.

Fwiw, I did a quick benchmark of the cost of adding a check for negative numbers, and it's pretty negligible (<2%) (the horizontal axis is `arr.len()`, and vertical axis is time to compute `arr.entropy()`; 'entropy2' is the checked implementation):

lines

```rust
    fn entropy2(&self) -> Option<A>
        where
            A: Float
    {
        if self.len() == 0 {
            None
        } else {
            let mut negative = false;
            let entropy = self.mapv(
                |x| {
                    if x < A::zero() {
                        negative = true;
                        A::zero()
                    } else if x == A::zero() {
                        A::zero()
                    } else {
                        x * x.ln()
                    }
                }
            ).sum();
            if negative {
                None
            } else {
                Some(-entropy)
            }
        }
    }
```

(Note that I wouldn't actually return an `Option` in this case; I'd return a `Result` instead.)

Of course, if we check for negative numbers, someone might wonder why we don't check for numbers greater than 1 too.

I could go either way on the negative number checks. The explicit check is nice, but it adds complexity and we aren't checking other things such as values greater than 1 or the sum of values not being 1, so I guess I'd lean towards leaving off negative number checks for right now.

> Thinking over it again, `Result<Option<A>, DimensionMismatch>` is probably the best signature, given that most of our methods return an `Option<A>` with `None` when arrays are empty.

Okay, that's fine.

> unless we want to use `DimensionMismatch::Empty` instead of `Option` in the rest of the API

That wouldn't be too bad, although I don't really like returning an error enum when only one of the variants is possible.

src/entropy.rs
jturner314: Yeah, `StatsBase.jl` is a good reference too. IMO it makes more sense to sum over all the axes (the current implementation of this PR) instead of just axis 0. If we wanted a function that sums over one axis, I'd suggest adding an `entropy_axis` or `axis_entropy` function.

src/entropy.rs
jturner314: Yeah, the issue is that `self.shape()` returns `&[usize]` which is a slice of unspecified length, so when you call `Array::zeros(anything.shape())`, you get a dynamic-dimensional array. (There have been a few issues about this in the `ndarray` repository from confused users, and it would be nice to resolve this raw edge of the API at some point.) `Zip` requires that all arrays have the same dimensionality, so trying to zip together an array with dimension `IxDyn` and arrays with dimension `D` doesn't work (unless you force `D = IxDyn`).

The issue thing is that calling [`ArrayBase::zeros()`](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.zeros) doesn't specify the storage of the array, so the compiler isn't sure what type of owned array to create (e.g. `Array` or `ArcArray`). Writing `Array::zeros()` tells the compiler that you specifically want an `Array` instance.

jturner314: I've created rust-ndarray/ndarray#591 to help improve the docs in this area.

LukeMathWalker:
I have addressed all issues I think - it should be ready for merging @jturner314

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
+//! Summary statistics (e.g. mean, variance, etc.).
```
jturner314:
```suggestion
//! Information theory (e.g. entropy, KL divergence, etc.).
```

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
...
+    ///
+    /// ```text
+    ///       n
+    /// S = - ∑ xᵢ ln(xᵢ)
+    ///      i=1
+    /// ```
+    ///
+    /// If the array is empty, `None` is returned.
+    ///
+    /// **Panics** if any element in the array is negative.
```
jturner314:
```suggestion
    /// **Panics** if `ln` of any element in the array panics (which can occur for negative values for some `A`).
```

Just a clarification that it won't always panic for negative numbers. (`ln` of negative `f32`/`f64` is `NaN`.)

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
...
+    /// ```text
+    ///              n
+    /// Dₖₗ(p,q) = - ∑ pᵢ ln(qᵢ/pᵢ)
+    ///             i=1
+    /// ```
+    ///
+    /// If the arrays are empty, Ok(`None`) is returned.
+    /// If the array shapes are not identical, `Err(ShapeMismatch)` is returned.
+    ///
+    /// **Panics** if for a  pair of elements *(pᵢ, qᵢ)* from *p* and *q* computing
```
jturner314:
```suggestion
    /// **Panics** if, for a pair of elements *(pᵢ, qᵢ)* from *p* and *q*, computing
```

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
...
+            .apply(|result, &p, &q| {
+                *result = {
+                    if p == A::zero() {
+                        A::zero()
+                    } else {
+                        p * (q / p).ln()
+                    }
+                }
+            });
+        let kl_divergence = temp.sum();
```
jturner314: This would be more clearly written as

```rust
let kl_divergence = -temp.sum();
Ok(Some(kl_divergence))
```

or

```rust
Ok(Some(-temp.sum()))
```

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
...
+            .apply(|result, &p, &q| {
+                *result = {
+                    if p == A::zero() {
+                        A::zero()
+                    } else {
+                        p * q.ln()
+                    }
+                }
+            });
+        let cross_entropy = temp.sum();
```
jturner314: Same comment here.

```rust
let cross_entropy = -temp.sum();
Ok(Some(cross_entropy))
```

or

```rust
Ok(Some(-temp.sum())
```

would be clearer.

src/entropy.rs
```diff
@@ -0,0 +1,407 @@
...
+            0.01398894, 0.01831756, 0.0227171 , 0.00736204, 0.01866295,
+        ];
+        // Computed using scipy.stats.entropy
+        let expected_entropy = 3.721606155686918;
+
+        assert_abs_diff_eq!(a.entropy().unwrap(), expected_entropy, epsilon = 1e-6);
+    }
+
+    #[test]
+    fn test_cross_entropy_and_kl_with_nan_values() -> Result<(), ShapeMismatch> {
```
jturner314: Woah, I didn't know that we can return `Result` from tests now. Cool!

src/errors.rs
```diff
@@ -0,0 +1,9 @@
+#[derive(Fail, Debug)]
```
jturner314: I strongly believe that we should implement `Error` instead of `Fail`, for a few reasons:

1. `failure` is ["experimental"](https://docs.rs/failure/0.1.5/failure/) and ["currently evolving"](https://github.com/rust-lang-nursery/failure#evolution), so while it may be fine for an end-user application, I don't think we should use it for a library like `ndarray-stats`.
2. The standard library's `Error` is being updated based on the `failure` experiment to resolve most of the issues with the old version of `Error`. ([RFC 2504](https://github.com/rust-lang/rfcs/blob/master/text/2504-fix-error.md), see esp. ["How this impacts failure"](https://github.com/rust-lang/rfcs/blob/master/text/2504-fix-error.md)) The [`.source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) method has been added, [`.cause()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.cause) has been deprecated, and [`.description()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.description) is "soft-deprecated".
3. It's annoying to have to call [`.compat()`](https://docs.rs/failure/0.1.5/failure/trait.Fail.html#method.compat) to get something that implements `Error`.

Implementing `Error` is pretty straightforward:

```rust
impl fmt::Display for ShapeMismatch {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "Array shapes do not match: {:?} and {:?}.", first_shape, second_shape)
    }
}

impl Error for ShapeMismatch {}
```

src/entropy.rs
LukeMathWalker: I thought it was just for `main`, but it turns out it works for tests as well :sunglasses:

src/errors.rs
LukeMathWalker: I wasn't aware of the RFC, and it also looks [pretty close to landing](https://github.com/rust-lang/rust/issues/53487). Given the "How this impacts failure" section I agree it makes sense to implement `Error` - I'll fix it straight-away.

LukeMathWalker:
All fixed, I'll squash and merge :+1:

Issue #16, opened 2018-11-19
Reduce cases where histogram strategies panic


jturner314:
See #13 for some discussion.

PR #25, opened 2019-02-02
Histogram error handling

LukeMathWalker:
Following on #16, it handles all current panic scenarios for our bin strategies, returning an `Option` when a suitable bin collection cannot be generated.

### DIFF
--- a/src/histogram/strategies.rs
+++ b/src/histogram/strategies.rs
@@ -211,7 +211,7 @@ where
             (Some(min), Some(max)) => {
                 let bin_width = compute_bin_width(min.clone(), max.clone(), n_bins);
                 let builder = EquiSpaced::new(bin_width, min.clone(), max.clone())?;
-                Self { builder }
+                Ok(Some(Self { builder }))
             }
             _ => Ok(None),
         }
@@ -255,7 +255,7 @@ where
             (Some(min), Some(max)) => {
                 let bin_width = compute_bin_width(min.clone(), max.clone(), n_bins);
                 let builder = EquiSpaced::new(bin_width, min.clone(), max.clone())?;
-                Self { builder }
+                Ok(Some(Self { builder }))
             }
             _ => Ok(None),
         }
@@ -299,9 +299,9 @@ where
             (Some(min), Some(max)) => {
                 let bin_width = compute_bin_width(min.clone(), max.clone(), n_bins);
                 let builder = EquiSpaced::new(bin_width, min.clone(), max.clone())?;
-                Self { builder }
+                Ok(Some(Self { builder }))
             }
-            _ => None,
+            _ => Ok(None),
         }
     }
 
@@ -339,7 +339,7 @@ where
     {
         let n_points = a.len();
         if n_points == 0 {
-            return None;
+            return Ok(None);
         }
 
         let mut a_copy = a.to_owned();
@@ -351,7 +351,7 @@ where
         match (a.min(), a.max()) {
             (Some(min), Some(max)) => {
                 let builder = EquiSpaced::new(bin_width, min.clone(), max.clone())?;
-                Self { builder }
+                Ok(Some(Self { builder }))
             }
             _ => Ok(None),
         }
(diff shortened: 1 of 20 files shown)


src/histogram/strategies.rs
```diff
@@ -143,11 +154,14 @@ impl<T> EquiSpaced<T>
     where
         T: Ord + Clone + FromPrimitive + NumOps + Zero
 {
-    /// **Panics** if `bin_width<=0`.
-    fn new(bin_width: T, min: T, max: T) -> Self
+    /// Returns `None` if `bin_width<=0` or `min` >= `max`.
+    fn new(bin_width: T, min: T, max: T) -> Option<Self>
```
munckymagik: Question: is returning _no object_ really a valid outcome or should it actually return a `Result<Self, WhateverError>` to signify a failure?

Hi, curious/lurking reviewer here. Saw the first "adventure" blog post and ended up here. Really great work that you people are doing and an inspiring article/tutorial.

If returning `None` is genuinely a valid, non-exceptional outcome, given certain args, then maybe a function name other than `new` would help clarify it's behavior? Just a suggestion.

LukeMathWalker: A `Result` is indeed a better return type there - I've been lazy because `EquiSpaced` is not part of the public API, but that's not a reason to be sloppy.

The creation method for the strategy is called `from_array`: do you think an `Option` is a good output type or would rather see a `Result` as a user?

munckymagik: I totally understand 👍 

As for the public `from_array` I'm hesitant to give a strong opinion yet, because I've only just started looking through the code. But the general rule of thumb I would apply is this:

* Use an `Option` if returning `None` means it's ok to continue
* If it's not ok to continue, the user may only see a problem later down the line, when a "symptomatic" error occurs rather than the actual cause.
* With `None` that symptom may just be no-output.

Is that any help?

LukeMathWalker: It totally is - going again through the code I think it makes sense to convert the output type to `Result`. I'll work on it :+1: 
Thanks for your input @munckymagik!

LukeMathWalker: Refactored to return `Result` - let me know what you think @munckymagik @jturner314

jturner314:
I've proposed some changes in LukeMathWalker/ndarray-stats#4. In #24, we decided to use `Result<Option<_>, _>` in some places for consistency with methods that return `None` for empty input arrays. After seeing how that approach affected this PR, though, I think we should use `Result<_, _>` instead after all. Doing so simplifies the implementation and public API. There really isn't much disadvantage to returning `Result` even for things like `.mean()` (which now returns `Result<A, EmptyInput>` instead of `Option<A>`). There should be no performance cost, and it's easy to convert `Result<T, E>` to `Option<T>` if needed with the `.ok()` method on `Result`. I think I've also addressed all of @munckymagik's comments.

LukeMathWalker:
Looking at the code in https://github.com/LukeMathWalker/ndarray-stats/pull/4 after the shift to `Result` everywhere I do agree it looks much cleaner (and less nested thanks to `?` and error conversions)!

munckymagik:
@jturner314 @LukeMathWalker looking really good.

There was just one other comment I made [about `|` vs `||`](https://github.com/jturner314/ndarray-stats/pull/25#discussion_r270620114).

Otherwise 🚢 it.

src/histogram/strategies.rs
```diff
@@ -141,13 +153,17 @@ impl<T> EquiSpaced<T>
...
-    fn new(bin_width: T, min: T, max: T) -> Self {
-        assert!(bin_width > T::zero());
-        Self {
-            bin_width,
-            min,
-            max,
+    /// Returns `Err(StrategyError)` if `bin_width<=0` or `min` >= `max`.
+    /// Returns `Ok(Self)` otherwise.
+    fn new(bin_width: T, min: T, max: T) -> Result<Self, StrategyError> {
+        if (bin_width <= T::zero()) | (min >= max) {
```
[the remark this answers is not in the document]
LukeMathWalker: Not deliberate, my bad :sweat_smile: Changed it to the lazy version :+1:

LukeMathWalker:
Taken care of that one @munckymagik :+1:

Issue #27, opened 2019-02-24
Reduce chance of overflow panics on Midpoint Interpolate


xd009642:
Midpoint interpolate is currently implemented like:
```Rust
let denom = T::from_u8(2).unwrap();
(lower.unwrap() + higher.unwrap()).mapv_into(|x| x / denom.clone())
``` 
This causes overflows most times I use it, whereas implementing like `(lower + (higher - lower))/2` would prevent this. 

Also, happy to do a PR for this if it would be helpful but I figured it was such a small change it might be quicker if you guys sorted it yourselves :+1:

xd009642:
I see the fix has been merged so I'll close this issue :+1:

PR #28, opened 2019-02-24
Fix overflow for Midpoint

LukeMathWalker:
This provides a fix for #27, I have also included a test to reproduce the issue so that we can be aware of regressions.
Unfortunately, it's a breaking change PR because I have had to add `Sub` to the trait bound. Should we just use `NumOps` for all `Interpolate` implementations @jturner314?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,8 +6,7 @@
 //! - [`partitioning`];
 //! - [`correlation analysis`] (covariance, pearson correlation);
 //! - [`histogram computation`].
-//
-//! !
+//!
 //! Please feel free to contribute new functionality! A roadmap can be found [`here`].
 //!
 //! Our work is inspired by other existing statistical packages such as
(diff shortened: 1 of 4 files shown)


LukeMathWalker:
Sorry, my wording was not precise. I meant using `NumOps` as trait bounds for all those requiring mathematical operations (mainly `Linear` and `Midpoint`). It's very easy to be minimal (`Add` but not `Sub`) but this leads to needless breaking changes in the API (such as this one) which could be probably be avoided by saying "you can use this if we can do arithmetics with your type".

jturner314:
Oh, I understand now. Sure, requiring `NumOps` for `Linear` and `Midpoint` seems reasonable and would help avoid breaking changes in the future.

By the way, thanks for taking care of this issue!

LukeMathWalker:
> By the way, thanks for taking care of this issue!

I am actually quite thrilled that this crate has its first legit user (and bug report, as it often goes) :rocket:

LukeMathWalker:
I have added the NumOps trait constraint - I'll squash and merge.

PR #29, opened 2019-02-26
Added a cumulative sum function to Histogram

xd009642:
This is equivalent to the cumsum for numpys histogram object. I've also added a basic test and a doc comment.

### DIFF
--- a/src/histogram/histograms.rs
+++ b/src/histogram/histograms.rs
@@ -78,6 +78,16 @@ impl<A: Ord> Histogram<A> {
     pub fn grid(&self) -> &Grid<A> {
         &self.grid
     }
+
+    /// Returns the cumulative distribution function of a histogram. 
+    /// Equivalent to the numpy histogram function cumsum
+    pub fn cumulative_sum(&self) -> ArrayD<usize> {
+        let mut total = 0;
+        self.counts.mapv(|x| {
+            total += x;
+            total
+        })
+    }
 }
 
 /// Extension trait for `ArrayBase` providing methods to compute histograms.
@@ -161,5 +171,25 @@ impl<A, S> HistogramExt<A, S> for ArrayBase<S, Ix2>
             let _ = histogram.add_observation(&point);
         }
         histogram
+    }
+}
+
+#[cfg(test)]
+mod auto_tests {
+    use super::*;
+    use histogram::histograms::HistogramExt;
+    use histogram::bins::{Bins, Edges};
+
+    #[test]
+    fn histogram_cdf() {
+        let data = arr2(&[[1], [2], [3], [4], [1], [1], [4], [5], [8], [8], [8]]);
+        let grid = Grid::from(vec![
+                              Bins::new(
+                                  Edges::from(vec![0,1,2,3,4,5,6,7,8,9]))]);
+
+        let histogram = data.histogram(grid);
+                       //0, 1  2  3  4  5  6  7  8
+        let cdf = arr1(&[0, 3, 4, 5, 7, 8, 8, 8, 11]).into_dyn();
+        assert_eq!(histogram.cumulative_sum(), cdf);
     }
 }
--- a/src/histogram/histograms.rs
+++ b/src/histogram/histograms.rs
@@ -185,7 +185,7 @@ mod auto_tests {
     use histogram::bins::{Bins, Edges};
 
     #[test]
-    fn histogram_cdf() {
+    fn histogram_cdf_1d() {
         let data = arr2(&[[1], [2], [3], [4], [1], [1], [4], [5], [8], [8], [8]]);
         let grid = Grid::from(vec![
                               Bins::new(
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
How do you handle grids with 2 or more dimensions? It's not clear in that case how to proceed with a cumulative sum.
This seems to only make sense for 1-dimensional histograms :thinking:

xd009642:
Let me go back to the drawing board codewise. I was however coming from the view of histogram equalisation in image processing where it will flatten the 2D image to get a 1D CDF.

xd009642:
Could add an `Option<Axis>` arg and do the CDF across one axis if present or for the flattened histogram if not present

jturner314:
Typically, the CDF for multiple variables is defined [like this](https://en.wikipedia.org/wiki/Cumulative_distribution_function#Multivariate_case). So, each element in the cumulative sum array would be

```
// 1D case
        i
c[i] =  ∑  a[m]
       m=0

// 2D case
          i   j
c[i,j] =  ∑   ∑  a[m,n]
         m=0 n=0

// 3D case
            i   j   k
c[i,j,k] =  ∑   ∑   ∑  a[m,n,o]
           m=0 n=0 o=0

// etc.
```

xd009642:
Okay I've implemented something I think should work. I'd like to get rid of the `to_owned` as that was really to defeat the borrow-checker and add a test case to ensure it actually works for multiple dimensions. But logically it looks sound to me and the previous test still passes

xd009642:
Okay I'm happy with this as an initial version if one of you guys wants to have a gander :smile:

LukeMathWalker:
Currently on holiday in a sunny place, I'll have a look once I come back :P

src/histogram/histograms.rs
```diff
@@ -78,6 +78,20 @@ impl<A: Ord> Histogram<A> {
     pub fn grid(&self) -> &Grid<A> {
         &self.grid
     }
+
+    /// Returns the cumulative distribution function of a histogram. 
+    /// Equivalent to the numpy histogram function cumsum
```
LukeMathWalker: I'd like to have a more detailed docstring here: it should include the math formula to compute the value indexed `i_1`, ..., `i_n` in the final array and a simple example using a 1d/2d array.

Given that you are mentioning NumPy, it would be good to link directly to the docs for `np.cumsum`.

src/histogram/histograms.rs
```diff
@@ -78,6 +78,20 @@ impl<A: Ord> Histogram<A> {
...
+    pub fn cumulative_sum(&self) -> ArrayD<usize> {
+        let mut cdf = self.counts.clone();
+        for i in 0..self.ndim() {
+            for j in 1..cdf.shape()[i] { 
+                let temp = cdf.index_axis(Axis(i), j - 1).to_owned();
```
LukeMathWalker: It's a shame that we have to use `to_owned` here, due to the borrow checker, even though the two slices are not-overlapping :disappointed:

LukeMathWalker:
The overall method is really a `np.cumsum` equivalent applied to the histogram matrix, so it kinda makes sense to have it as a `cumsum` method on `ArrayBase` instead of special-casing it for histograms. 
Is this something `ndarray` might want to host @jturner314 or should we add it to one of `ndarray-stats` extension traits?

Irrespectively of where it is going to end up, it would be worth it to implement it as a free function taking an `ArrayBase` as input - then we can place it where it looks more convenient API-wise.

I have left some other minor comments @xd009642, but it looks good to me overall!

src/histogram/histograms.rs
LukeMathWalker: Actually, this can be avoided using `split_at`!

xd009642: So I just started trying to write that but found myself with two matrices of different sizes and thinking I'd like to get the last element of one for a given axis and the first of another and it felt too complicated.. But I might just not be picturing it in the right way :confused:

jturner314: The simplest approach would be to avoid the issue by using `.lanes_mut()`; something like this [*untested*]:

```rust
let mut cdf = self.counts.clone();
for ax in 0..cdf.ndim() {
    for lane in cdf.lanes_mut(Axis(ax)) {
        for i in 1..lane.len() {
            lane[i] += lane[i-1];
        }
    }
}
```

jturner314:
Yes, I think it makes sense to add `cumsum` and `cumsum_axis` to `ndarray`. That would also make it easier to justify additional code for optimizing iteration order and dealing with numerical precision issues (for floating point array elements).

xd009642:
Okay I'll transfer my work to ndarray and open a PR there :+1:

src/histogram/histograms.rs
LukeMathWalker: If that works @jturner314, it's surely cleaner. Unfortunately it kind of goes the way you described @xd009642 using `split_at` :thinking: You need to split at `j` and then grab the last lane of the first part and the first lane of the second part.

jturner314: I just tested it, and my example does work (with the addition of `mut` before `lane`). It does rely on the element type being `Copy`, though; for general `A` you'd run into the same issue.

Fwiw, regarding the inconvenience of a solution based on `.split_at()`, the next release of `ndarray` will have a [`multislice!`](https://github.com/rust-ndarray/ndarray/blob/master/src/slice.rs#L653-L739) macro that makes it possible to cleanly take multiple disjoint, mutable slices simultaneously. This PR makes me realize that we should add a `multislice_axis!` macro too.

xd009642:
@jturner314 I was looking at ndarray PRs and saw this one https://github.com/rust-ndarray/ndarray/pull/513 it seems that it would probably be better to get that PR finished and merged in than start a new one that's less generic. Although I can continue with one since that one seems to have stalled

LukeMathWalker:
I think it makes sense to file your PR in any case @xd009642  - when https://github.com/rust-ndarray/ndarray/pull/513 lands/resumes activity we will take care of rephrasing `cumsum` in those terms, if possible.

PR #30, opened 2019-03-09
Implement argmin argmax

phungleson:
Hey mates, this PR currently has a rough `impl` for `argmin` for discussions.

Please have some comments. Once everyone is happy with the implementation, I will add necessary docs, tests.. and other methods.

### DIFF
--- a/src/quantile.rs
+++ b/src/quantile.rs
@@ -182,6 +182,13 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
+    /// Finds the indices of the minimum values along an axis.
+    ///
+    /// Returns `None` if the array is empty.
+    fn argmin(&self) -> Option<Vec<D::Pattern>>
+    where
+        A: PartialOrd;
+
     /// Finds the elementwise minimum of the array.
     ///
     /// Returns `None` if any of the pairwise orderings tested by the function
@@ -278,6 +285,22 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
+    fn argmin(&self) -> Option<Vec<D::Pattern>>
+    where
+        A: PartialOrd,
+    {
+        let min = self.min()?;
+        let mut args = Vec::<D::Pattern>::new();
+
+        self.indexed_iter().for_each(|(pattern, elem)| {
+            if elem.partial_cmp(min) == Some(cmp::Ordering::Equal) {
+                args.push(pattern)
+            }
+        });
+
+        Some(args)
+    }
+
     fn min(&self) -> Option<&A>
     where
         A: PartialOrd,
--- a/src/quantile.rs
+++ b/src/quantile.rs
@@ -182,14 +182,14 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
-    /// Finds the indices of the minimum values of the array.
+    /// Finds the first index of the minimum value of the array.
     ///
     /// Returns `None` if any of the pairwise orderings tested by the function
     /// are undefined. (For example, this occurs if there are any
     /// floating-point NaN values in the array.)
     ///
     /// Returns `None` if the array is empty.
-    fn argmin(&self) -> Option<Vec<D::Pattern>>
+    fn argmin(&self) -> Option<D::Pattern>
     where
         A: PartialOrd;
 
@@ -214,14 +214,14 @@ where
         A: MaybeNan,
         A::NotNan: Ord;
 
-    /// Finds the indices of the maximum values of the array.
+    /// Finds the first index of the maximum value of the array.
     ///
     /// Returns `None` if any of the pairwise orderings tested by the function
     /// are undefined. (For example, this occurs if there are any
     /// floating-point NaN values in the array.)
     ///
     /// Returns `None` if the array is empty.
-    fn argmax(&self) -> Option<Vec<D::Pattern>>
+    fn argmax(&self) -> Option<D::Pattern>
     where
         A: PartialOrd;
 
@@ -300,20 +300,25 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
-    fn argmin(&self) -> Option<Vec<D::Pattern>>
+    fn argmin(&self) -> Option<D::Pattern>
     where
         A: PartialOrd,
     {
-        let min = self.min()?;
-        let mut args = Vec::<D::Pattern>::new();
+        let min = self.first()?;
+        let mut pattern_min = self.dim();
 
-        self.indexed_iter().for_each(|(pattern, elem)| {
-            if elem.partial_cmp(min) == Some(cmp::Ordering::Equal) {
-                args.push(pattern)
-            }
-        });
+        self.indexed_iter()
+            .fold(Some(min), |acc, (pattern, elem)| {
+                match elem.partial_cmp(acc?)? {
+                    cmp::Ordering::Less => {
+                        pattern_min = pattern;
+                        Some(elem)
+                    }
+                    _ => acc,
+                }
+            })?;
 
-        Some(args)
+        Some(pattern_min)
     }
 
     fn min(&self) -> Option<&A>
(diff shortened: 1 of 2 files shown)


munckymagik:
Hi @phungleson thanks for the contribution!

I think it was a great idea to raise this PR and get a conversation started 👍. I'm a volunteer. I am happy to keep an eye out for comments and changes, review/discuss ideas and generally cheer you on.

It certainly looks like this code would work, but I agree we need to get to a point where we are all happy it's the right implementation.

I suggest the next step in achieving that is to wrap some tests around it so we can experiment with the implementation without risk of breaking things. May not need to be deep coverage at this stage, but it's up to you. I would think about relevant ndarray shapes, types and error cases.

From quick review it looks like its run-time complexity is `O(2n)` big-oh of 2n. If that's the case, then my instinct is we could improve on that.

Have you had a look for ideas from existing implementations in other technologies? For example numpy and Julia - how do they implement this? Are there any special algorithms people use that are better than the intuitive approach?

Thanks. I hope this helps.

phungleson:
Hey @munckymagik thanks for the comment! :) I have added a few tests and `argmax`

For `numpy.argmin`, it shows that it delegates to some sort of `getattr`

https://github.com/numpy/numpy/blob/v1.10.1/numpy/core/fromnumeric.py#L967
https://github.com/numpy/numpy/blob/v1.10.1/numpy/core/fromnumeric.py#L43-L53

For Julia, it seems like it just get the first index of min and returns that value (not a `vec`).

https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/base/array.jl#L2159
https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/base/array.jl#L2096-L2115

Perhaps we can do it in a way similar to Julia but we can return a `vec` of indices of min value. What do you think?

LukeMathWalker:
Thanks for jumping in @munckymagik! The issues and PRs of this repo are getting crowded, that's super cool :rocket: 

I guess the main choice we have to make concerns the number of outputs to be returned: if there is more than one minimum/maximum value, do we return just the indices for the first minimum/maximum value we encounter or do we want to return them all?

If we align with Julia/NumPy, returning just the first one, then we should be able to get complexity down to `O(n)` - you don't need to "precompute" the minimum using the `min` method :+1:

src/quantile.rs
```diff
@@ -8,7 +8,7 @@ use {MaybeNan, MaybeNanExt, Sort1dExt};
 pub mod interpolate {
     use ndarray::azip;
     use ndarray::prelude::*;
-    use num_traits::{FromPrimitive, ToPrimitive, NumOps};
+    use num_traits::{FromPrimitive, NumOps, ToPrimitive};
```
phungleson: This is probably due to `rustfmt`, I can remove if it is un-necessary.

LukeMathWalker: No worries, I am working on getting `rustfmt` in the CI pipeline :+1: (see #31)

phungleson:
Cool thanks, I have changed to return a single value. As the idea of this lib is to be closer to `numpy`, perhaps we can keep this?

Let me know if returning a `vec` is desirable, it shouldn't be too hard.

src/quantile.rs
```diff
@@ -278,6 +300,27 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
+    fn argmin(&self) -> Option<D::Pattern>
+    where
+        A: PartialOrd,
+    {
+        let min = self.first()?;
+        let mut pattern_min = self.dim();
```
phungleson: This is not correct do to my limited knowledge of `ndarray`, how do I get the pattern of the first element?

src/quantile.rs
```diff
@@ -278,6 +300,27 @@ where
...
 {
+    fn argmin(&self) -> Option<D::Pattern>
+    where
+        A: PartialOrd,
+    {
+        let min = self.first()?;
+        let mut pattern_min = D::zeros(self.ndim()).into_pattern();
+
+        self.indexed_iter()
+            .fold(Some(min), |acc, (pattern, elem)| {
```
LukeMathWalker: I think this should still work if we remove `Some` here - in this way we can get rid of the internal `?` operator on `acc` when calling `partial_cmp`.

LukeMathWalker: A stylistic suggestion: renaming `acc` to `current_min` (or something similar) would slightly improve readability.

src/quantile.rs
```diff
@@ -278,6 +300,27 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
+    fn argmin(&self) -> Option<D::Pattern>
+    where
+        A: PartialOrd,
+    {
+        let min = self.first()?;
```
LukeMathWalker: A stylistic suggestion: `min` is not the variable we use to keep track of the minimum - I'd rename it to something closer to what it does/is (`first_value`?).

tests/quantile.rs
```diff
@@ -5,10 +5,21 @@ extern crate ndarray_stats;
...
+fn test_argmin() {
+    let a = array![[0, 5, 3], [2, 0, 6]];
+    assert_eq!(a.argmin(), Some((0, 0)));
+
+    let a = array![[1., 5., 3.], [2., 0., 6.]];
+    assert_eq!(a.argmin(), Some((1, 1)));
+
+    let a = array![[1., 5., 3.], [2., ::std::f64::NAN, 6.]];
+    assert_eq!(a.argmin(), None);
+}
```
LukeMathWalker: We should add a test case, both for `argmin` and `argmax`, taking an empty array as input.

tests/quantile.rs
```diff
@@ -36,6 +47,18 @@ fn test_min_skipnan_all_nan() {
     assert!(a.min_skipnan().is_nan());
 }
 
+#[test]
+fn test_argmax() {
```
LukeMathWalker: We should test what happens when there is more than one maximum value in the array - you have it of `argmin`, but it's missing for `argmax`.

src/quantile.rs
```diff
@@ -182,6 +182,17 @@ where
     S: Data<Elem = A>,
     D: Dimension,
 {
+    /// Finds the first index of the minimum value of the array.
+    ///
+    /// Returns `None` if any of the pairwise orderings tested by the function
+    /// are undefined. (For example, this occurs if there are any
+    /// floating-point NaN values in the array.)
+    ///
+    /// Returns `None` if the array is empty.
```
LukeMathWalker: Can we add a doc test example to both argmin/argmax? `D::Pattern` is the right signature, but it might feel a bit obscure to a new comer looking at the docs for the first time (even more if they don't know `ndarray` and they have to dig) - a simple example with a 2d array showing that it's just a `(usize, usize)` would go a long way to make it more obvious :pray:

LukeMathWalker:
I left some comments for polishing, but the functionality is there :100: @phungleson

src/quantile.rs
phungleson: Hey mate, `Some` cannot be easily removed, as `partial_cmp` can be `None` and we want to return `None` for such case.

BTW, would old `for` loop makes thing easier to read?

```suggestion
let mut current_min = self.first()?;
let mut current_pattern_min = D::zeros(self.ndim()).into_pattern();

for (pattern, elem) in self.indexed_iter() {
    if elem.partial_cmp(current_min)? == cmp::Ordering::Less {
        current_pattern_min = pattern;
        current_min = elem
    }
}

Some(current_pattern_min)
```

src/quantile.rs
LukeMathWalker: #31 has been merged, this should be ok now once you merge master back into your branch :+1:

src/quantile.rs
LukeMathWalker: You are right, my bad :sweat_smile: 
The `for` loop version looks cleaner to me as well, let's go with that!

phungleson:
@LukeMathWalker Hey mate, I have made the changes, let me know if anything else is still missing.

I probably will add `argmin_skipnan` `argmax_skipnan` in a separate PR to keep this one short.

LukeMathWalker:
I agree with small PRs - feel free to work the `skipnan` versions on a separate one.
This one looks good to me - I'll squash and merge.

PR #31, opened 2019-03-09
Add rustfmt to the CI pipeline

LukeMathWalker:
This should avoid spurious diffs in PRs due to IDE's settings and/or style nitpicks.


LukeMathWalker:
It's complaining because it touches code that is not tested, but it's actually fine CI-wise. @jturner314

jturner314:
Some of the formatting is a little ugly (primarily unnecessary line breaks in the `array!` and `assert_eq!` macro), but it's worth the convenience of being able to just call `rustfmt` during development without worrying about accidentally reformatting existing code. Thanks!

PR #40, opened 2019-04-11
Update to Rust 2018 and replace quickcheck! with #[quickcheck]

jturner314:


jturner314:
I've also bumped the Rust version to 1.34.

PR #42, opened 2019-04-19
Add link to #rust-sci in README

jturner314:
It's helpful to have a place for users to chat. The link is to #rust-sci in the Mibbit client (the same client used for the "Mozilla IRC" link on https://www.rust-lang.org/community). Mibbit is nice because it provides a way for people to join the chat using their browser without even having to register.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Dependencies status](https://deps.rs/repo/github/jturner314/ndarray-stats/status.svg)](https://deps.rs/repo/github/jturner314/ndarray-stats)
 [![Crate](https://img.shields.io/crates/v/ndarray-stats.svg)](https://crates.io/crates/ndarray-stats)
 [![Documentation](https://docs.rs/ndarray-stats/badge.svg)](https://docs.rs/ndarray-stats)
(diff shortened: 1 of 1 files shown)


munckymagik:
Seeing as [Moz IRC is going away](https://blog.rust-lang.org/2019/04/26/Mozilla-IRC-Sunset-and-the-Rust-Channel.html) shall we direct chat to a channel within Discord or similar?

LukeMathWalker:
We either point to Discord or to the new channel that some of the users of IRC are creating - do you have more details here @jturner314? I haven't been following the discussion closely.

jturner314:
The #rust-sci IRC channel is moving to oftc.net. (Web chat: https://webchat.oftc.net/?channels=rust-sci) (If you're using Matrix/Riot, the OFTC bridge is `@oftc-irc:matrix.org` and the nickserv is `@_oftc_NickServ:matrix.org`.) A few people have moved over, but it's very quiet.

I'd like to include a link to #rust-sci because that's where I'm available, and it'll be convenient for users who are familiar with IRC and/or unable to use Discord. However, it might be a good idea to link to the Discord channel too because the IRC channel is so quiet. (I'm assuming there's more activity on Discord? I'm not able to access it.)

PR #43, opened 2019-05-01
Add standard deviation

oberien:


LukeMathWalker:
Hi @oberien!

Thanks for the taking the time to draft a PR.
Right now there is a method in `ndarray` to compute the standard deviation along an axis ([`std_axis`](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.std_axis)).
Should we add this method here @jturner314 or should we rather move it upstream in `ndarray`?

jturner314:
`ndarray` has `sum`, `sum_axis`, `mean`, `mean_axis`, `std_axis`, and `var_axis`. I think we should add `std` and `var` to `ndarray` for consistency, ideally using the same algorithm as the existing `std_axis` and `var_axis` methods and including the same `ddof` parameter.

munckymagik:
Did this bit of work move to `ndarray` then? If so shall we close this to clear the decks?

LukeMathWalker:
I have opened an issue in `ndarray` to keep track of it ( https://github.com/rust-ndarray/ndarray/issues/655 ) - we can close here.

PR #46, opened 2019-05-10
Improve coverage testing

jturner314:
The `tarpaulin` README says "Because of the use of nightly proc-macro features you'll need to reinstall tarpaulin each time unless you're keeping to a specific nightly version. If you are keeping to a specific nightly you can remove the `-f` flag in the example travis file."

We haven't pinned a nightly version, so we should use the `-f` option.


jturner314:
I've just added another commit that enables doc tests in coverage testing. (This is a feature new in `tarpaulin` 0.8.0.)

PR #48, opened 2019-08-06
Deviation - Allow mixed ownership

LukeMathWalker:
Small change to allow deviation measures to be computed for pairs of arrays with different ownership over their data.

This snippet
```rust
let a = array![0., 0.];
let b = array![1., 0.];

assert_eq!(a.count_eq(&b.view())?, 1);
```
does not compile on master, but it does on this branch.

I bumped on this limitation while working on a k-means example (https://github.com/rust-ndarray/ndarray-examples/commit/4e4692af00d86eda5f0e2f731a6062689977d7a9 - `find_closest_centroid`).

### DIFF
--- a/tests/deviation.rs
+++ b/tests/deviation.rs
@@ -12,12 +12,12 @@ use std::f64;
 fn test_count_eq() -> Result<(), MultiInputError> {
     let a = array![0., 0.];
     let b = array![1., 0.];
-    let c = array![0., 1.];
+    let mut c = array![0., 1.];
     let d = array![1., 1.];
 
     assert_eq!(a.count_eq(&a)?, 2);
-    assert_eq!(a.count_eq(&b)?, 1);
-    assert_eq!(a.count_eq(&c)?, 1);
+    assert_eq!(a.count_eq(&b.view())?, 1);
+    assert_eq!(a.count_eq(&c.view_mut())?, 1);
     assert_eq!(a.count_eq(&d)?, 0);
 
     Ok(())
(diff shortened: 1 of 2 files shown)


tests/deviation.rs
```diff
@@ -12,12 +12,12 @@ use std::f64;
 fn test_count_eq() -> Result<(), MultiInputError> {
     let a = array![0., 0.];
     let b = array![1., 0.];
-    let c = array![0., 1.];
+    let mut c = array![0., 1.];
```
munckymagik: Suggestion: create a dedicated test case for ownership variation.

The individual tests for each function, towards top of this file, focus on the desired behaviour mathematically. Towards the end of the file the tests focus on shared issues which make them a consistently behaving module. 

So my suggestion is to create a test focused on the issue of ownership variation because it's a quality that all the function in this module need to share.

LukeMathWalker: You always get me when I am being lazy. I'll refactor :crossed_fingers:

LukeMathWalker:
> This is great. This is a exactly the kind of feedback I hoped we'd get from writing some examples.

100% agree - that's why I loved your idea :grin: 

> Great job getting that process started btw, I'm afraid life has been a bit busy for me in recent weeks and I haven't had head space to get involved yet.

No worries at all, I went through a similar period a couple of months ago. Just take care of yourself :+1:

tests/deviation.rs
LukeMathWalker: Done - check it out @munckymagik :+1:

PR #51, opened 2019-09-16
Weighted mean

nilgoyette:
Here's a first version of `weighted_mean` and `weighted_mean_axis`. 

Disclaimers:

1. I don't really know where `weighted_mean` and friends  should go. Is `summary_statistics` ok?
2. I had to move `return_err_if_empty` and `return_err_unless_same_shape` because ther are useful elsewhere.
3. There's a little code-copy in `weighted_mean_axis`, to avoid `(2 conditions + 1 unwrap) x nb_lanes`. Maybe I could create an inner function called `inner_weighted_mean` or something, then call it in both functions?

Questions:

1. Why are the `summary_statistics` tests (and others) not in `/tests/*`? I thought that the public API was supposed to be tested outside the crate. Is this not a "standard"?

### DIFF
--- a/src/summary_statistics/means.rs
+++ b/src/summary_statistics/means.rs
@@ -24,7 +24,15 @@ where
         }
     }
 
-    fn weighted_mean(&self, weights: &ArrayBase<S, D>) -> Result<A, MultiInputError>
+    fn weighted_mean(&self, weights: &Self) -> Result<A, MultiInputError>
+    where
+        A: Copy + Div<Output = A> + Mul<Output = A> + Zero
+    {
+        let weighted_sum = self.weighted_sum(weights)?;
+        Ok(weighted_sum / weights.sum())
+    }
+
+    fn weighted_sum(&self, weights: &ArrayBase<S, D>) -> Result<A, MultiInputError>
     where
         A: Copy + Mul<Output = A> + Zero,
     {
@@ -42,6 +50,20 @@ where
         weights: &ArrayBase<S, Ix1>,
     ) -> Result<Array<A, D::Smaller>, MultiInputError>
     where
+        A: Copy + Div<Output = A> + Mul<Output = A> + Zero,
+        D: RemoveAxis
+    {
+        let mut weighted_sum = self.weighted_sum_axis(axis, weights)?;
+        weighted_sum.mapv_inplace(|v| v / weights.sum());
+        Ok(weighted_sum)
+    }
+
+    fn weighted_sum_axis(
+        &self,
+        axis: Axis,
+        weights: &ArrayBase<S, Ix1>,
+    ) -> Result<Array<A, D::Smaller>, MultiInputError>
+    where
         A: Copy + Mul<Output = A> + Zero,
         D: RemoveAxis,
     {
@@ -53,7 +75,7 @@ where
             }));
         }
 
-        // We could use `lane.weighted_mean` here, but we're avoiding 2
+        // We could use `lane.weighted_sum` here, but we're avoiding 2
         // conditions and an unwrap per lane.
         Ok(self.map_axis(axis, |lane| {
             lane.iter()
@@ -239,9 +261,9 @@ mod tests {
     fn test_means_with_nan_values() {
         let a = array![f64::NAN, 1.];
         assert!(a.mean().unwrap().is_nan());
-        assert!(a.weighted_mean(&array![1.0, f64::NAN]).unwrap().is_nan());
+        assert!(a.weighted_sum(&array![1.0, f64::NAN]).unwrap().is_nan());
         assert!(a
-            .weighted_mean_axis(Axis(0), &array![1.0, f64::NAN])
+            .weighted_sum_axis(Axis(0), &array![1.0, f64::NAN])
             .unwrap()
             .into_scalar()
             .is_nan());
@@ -254,11 +276,11 @@ mod tests {
         let a: Array1<f64> = array![];
         assert_eq!(a.mean(), Err(EmptyInput));
         assert_eq!(
-            a.weighted_mean(&array![1.0]),
+            a.weighted_sum(&array![1.0]),
             Err(MultiInputError::EmptyInput)
         );
         assert_eq!(
-            a.weighted_mean_axis(Axis(0), &array![1.0]),
+            a.weighted_sum_axis(Axis(0), &array![1.0]),
             Err(MultiInputError::EmptyInput)
         );
         assert_eq!(a.harmonic_mean(), Err(EmptyInput));
@@ -269,9 +291,9 @@ mod tests {
     fn test_means_with_empty_array_of_noisy_floats() {
         let a: Array1<N64> = array![];
         assert_eq!(a.mean(), Err(EmptyInput));
-        assert_eq!(a.weighted_mean(&array![]), Err(MultiInputError::EmptyInput));
+        assert_eq!(a.weighted_sum(&array![]), Err(MultiInputError::EmptyInput));
         assert_eq!(
-            a.weighted_mean_axis(Axis(0), &array![]),
+            a.weighted_sum_axis(Axis(0), &array![]),
             Err(MultiInputError::EmptyInput)
         );
         assert_eq!(a.harmonic_mean(), Err(EmptyInput));
@@ -312,7 +334,7 @@ mod tests {
         // weighted_mean with itself, normalized
         let weights = &a / a.sum();
         assert_abs_diff_eq!(
-            a.weighted_mean(&weights).unwrap(),
+            a.weighted_sum(&weights).unwrap(),
             expected_weighted_mean,
             epsilon = 1e-12
         );
@@ -320,7 +342,7 @@ mod tests {
         let data = a.into_shape((2, 5, 5)).unwrap();
         let weights = array![0.1, 0.3, 0.25, 0.15, 0.2];
         assert!(data
-            .weighted_mean_axis(Axis(1), &weights)
+            .weighted_sum_axis(Axis(1), &weights)
             .unwrap()
             .all_close(
                 &array![
@@ -330,7 +352,7 @@ mod tests {
                 1e-8
             ));
         assert!(data
-            .weighted_mean_axis(Axis(2), &weights)
+            .weighted_sum_axis(Axis(2), &weights)
             .unwrap()
             .all_close(
                 &array![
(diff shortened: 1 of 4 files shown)


fmorency:
I think we should mimic the numpy behavior here and not assume the weights are pre-normalized. I think we should divide by the weight sum inside the `weighted_mean` method.

src/summary_statistics/means.rs
```diff
@@ -250,6 +307,48 @@ mod tests {
             expected_geometric_mean,
             epsilon = 1e-12
         );
+
+        // Same as mean
+        let weights = Array1::from_elem(a.len(), 1.0 / a.len() as f64);
```
LukeMathWalker: It would be really cool to have both this check and the one below as standalone tests, ideally using `quickcheck`.

LukeMathWalker:
> 1. I don't really know where `weighted_mean` and friends  should go. Is `summary_statistics` ok?

I'd say it makes sense for them to go there, given that it's where `mean` is.

> 2. I had to move `return_err_if_empty` and `return_err_unless_same_shape` because ther are useful elsewhere.

It makes perfect sense.

> 3. There's a little code-copy in `weighted_mean_axis`, to avoid `(2 conditions + 1 unwrap) x nb_lanes`. Maybe I could create an inner function called `inner_weighted_mean` or something, then call it in both functions?

I don't feel too strongly in either direction - if you feel like doing it, all the better :+1: 

> Questions:
> 
> 1. Why are the `summary_statistics` tests (and others) not in `/tests/*`? I thought that the public API was supposed to be tested outside the crate. Is this not a "standard"?

I'd say that it's standard for integration tests to go outside the crate, yes. On the other side, it's sometimes nice to have tests next to the code they refer to - we haven't been very consistent across the crate. It would probably makes sense to migrate them to the `tests` folder.

LukeMathWalker:
> I think we should mimic the numpy behavior here and not assume the weights are pre-normalized. I think we should divide by the weight sum inside the `weighted_mean` method.

I am not sure - I am concerned about introducing rounding errors doing a normalization step that might or might not be necessary.
I would lean on the side of being explicit and letting the user taking care of normalisation, if they need it.

fmorency:
> > I think we should mimic the numpy behavior here and not assume the weights are pre-normalized. I think we should divide by the weight sum inside the `weighted_mean` method.
> 
> I am not sure - I am concerned about introducing rounding errors doing a normalization step that might or might not be necessary.
> I would lean on the side of being explicit and letting the user taking care of normalisation, if they need it.

I see your point and I tend to agree. 

However, my concern is that the `weighted_mean` function doesn't return an actual weighted mean, except then the weights are normalized. From the definition in the documentation, I would tend to
  - Rename this function to `weighted_sum` and note in the documentation that if the weights are pre-normalized, the result is a weighted mean
  - Write an additional function `weighted_mean` that actually computes the weighted mean given any input - ie. divide by the sum of the weights.

Thoughts?

LukeMathWalker:
> > > I think we should mimic the numpy behavior here and not assume the weights are pre-normalized. I think we should divide by the weight sum inside the `weighted_mean` method.
> > 
> > 
> > I am not sure - I am concerned about introducing rounding errors doing a normalization step that might or might not be necessary.
> > I would lean on the side of being explicit and letting the user taking care of normalisation, if they need it.
> 
> I see your point and I tend to agree.
> 
> However, my concern is that the `weighted_mean` function doesn't return an actual weighted mean, except then the weights are normalized. From the definition in the documentation, I would tend to
> 
> * Rename this function to `weighted_sum` and note in the documentation that if the weights are pre-normalized, the result is a weighted mean
> * Write an additional function `weighted_mean` that actually computes the weighted mean given any input - ie. divide by the sum of the weights.
> 
> Thoughts?

I think that makes perfect sense :+1:

src/summary_statistics/means.rs
nilgoyette: I tried fixing this, but it's less "sexy" than I thought. Mostly because of the minimum size. I do agree that the tests are better now though.

src/summary_statistics/mod.rs
```diff
@@ -45,16 +49,32 @@ where
     /// [`arithmetic weighted mean`] https://en.wikipedia.org/wiki/Weighted_arithmetic_mean
     fn weighted_mean(&self, weights: &Self) -> Result<A, MultiInputError>
     where
-        A: Copy + Mul<Output = A> + Zero;
+        A: Copy + Div<Output = A> + Mul<Output = A> + Zero;
 
-    /// Returns the [`arithmetic weighted mean`] x̅ along `axis`. Assumes that the weights are
-    /// already normalized.
+    /// Like `weighted_mean`, but assumes that the `weights` are normalized. In that case, this
```
LukeMathWalker: I would just describe what a weighted sum is and then point to the equivalence with weighted_mean if the weights are normalised.

src/summary_statistics/mod.rs
```diff
@@ -28,13 +28,17 @@ where
     where
         A: Clone + FromPrimitive + Add<Output = A> + Div<Output = A> + Zero;
 
-    /// Returns the [`arithmetic weighted mean`] x̅ of all elements in the array. Assumes that the
-    /// weights are already normalized.
+    /// Returns the [`arithmetic weighted mean`] x̅ of all elements in the array. Use `weighted_sum`
```
LukeMathWalker: What happens if the weights sum to `0`? I assume the function panics:
- we should add it to the docs;
- we should add a test for it.

nilgoyette: Oh, I forgot about this. Thank you.

nilgoyette: In fact, it panics on integer and it returns a NaN on floating point, which is quite normal. Do we want to panic on floating point numbers or we let the users learn the hard facts of IEEE754?

LukeMathWalker: I would let them know the hard facts, unfortunately :joy: 
We could use something along the same lines of what we say for `mean_axis` in `ndarray`: 

> Panics if axis is out of bounds or if the length of the axis is zero and division by zero panics for type A.

Which makes me think that we need to add a test and a note about zero-length axis on both `weighted_mean_axis` and `weighted_sum_axis`.

nilgoyette: Sorry, I don't understand this part. How can "the length of the axis" be zero? Are we talking about a `Array0` (single number)? Surely the len of an `Array0` is 1 and not 0? Or it's considered a data **point** and it has no length?

Also, reading `ndarray`'s version, it seems like it's saying "Panics if x OR (y AND z)" Should it be all "or"?

nilgoyette: Sorry to disturb you again, but I'm lost here.

1. Those zero-length dimensions don't make sense to me. It looks like they are uselessly complex empty arrays. Do you have more information on them? Because I can't code something for a "feature" that I don't understand.
2. I removed the `* MultiInputError::EmptyInput if self is empty` error on `weighted_mean_axis` and `weighted_sum_axis` because I don't understand why they should return an error when the array is empty. It does seem to make sense for `weighted_mean_axis` because we divide all lanes by weights_sum, but there's no lane to divide anyway.
3. I can't respect `Panics if axis is out of bounds or if the length of the axis is zero and division by zero panics for type A.` because my functions use `map_axis`, which panics with an error not defined in [`map_axis` documentation](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.map_axis).
```
let a = Array2::<f32>::zeros((0, 20));
a.mean_axis(Axis(0)); // pass
a.weighted_mean_axis(Axis(0), &Array1::zeros(0)).unwrap(); // fail
    panicked at 'collapse_axis: Index 0 must be less than axis length 0 for array with shape [0, 20]
```

Of course I could try to use something else (there are other ways to code `weighted_sum_axis`), but I don't see why I should, mostly because of 1).

LukeMathWalker: No issue at all, happy to reach consensus on these changes :smile: 

> 1. Those zero-length dimensions don't make sense to me. It looks like they are uselessly complex empty arrays. Do you have more information on them? Because I can't code something for a "feature" that I don't understand.

I wouldn't indeed consider them a feature - they are edge cases (empty arrays with associated non-trivial dimension-related information), but given that we are writing a library we need to take care of them (or at least document what happens if you pass them in).
Would it possible to forbid them? It's a possibility, but as long as `ndarray`'s codebase allows them, we need to deal with them. 

> 2. I removed the `* MultiInputError::EmptyInput if self is empty` error on `weighted_mean_axis` and `weighted_sum_axis` because I don't understand why they should return an error when the array is empty. It does seem to make sense for `weighted_mean_axis` because we divide all lanes by weights_sum, but there's no lane to divide anyway.

I agree on `weighted_sum_axis` / `weighted_sum` - there is a very good default value to return (`A::zero()`) and the behaviour is consistent with [`sum`](https://docs.rs/ndarray/0.13.0/ndarray/struct.ArrayBase.html#method.sum) in `ndarray`.
For the mean though, it gets trickier. What is the mean of an empty array of integers? `0`? If would argue that there is no sound answer to that question, hence it makes sense to return `None`. Given that those functions can fail in multiple ways (mismatched shapes), it's more ergonomic to return it as an error variant than a `Result<Option<A>, ErrType>` (see [`mean`](https://docs.rs/ndarray/0.13.0/ndarray/struct.ArrayBase.html#method.mean) in `ndarray`).

> 3. I can't respect `Panics if axis is out of bounds or if the length of the axis is zero and division by zero panics for type A.` because my functions use `map_axis`, which panics with an error not defined in [`map_axis` documentation](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.map_axis).
> 
> ```
> let a = Array2::<f32>::zeros((0, 20));
> a.mean_axis(Axis(0)); // pass
> a.weighted_mean_axis(Axis(0), &Array1::zeros(0)).unwrap(); // fail
>     panicked at 'collapse_axis: Index 0 must be less than axis length 0 for array with shape [0, 20]
> ```
> 
> Of course I could try to use something else (there are other ways to code `weighted_sum_axis`), but I don't see why I should, mostly because of 1).

This was a bug in `ndarray:0.12.1` (indeed when dealing with `0`-length axis :stuck_out_tongue:). It has been fixed in `0.13` - as soon as the CI is finished in https://github.com/rust-ndarray/ndarray-stats/pull/52 I will merge it into `master`, so that you can get the fix in this branch as well.

LukeMathWalker: https://github.com/rust-ndarray/ndarray-stats/pull/52 has been merged - I'd say that we could this PR merged and then cut a release :+1:

nilgoyette:
I think it's better now. I had to remove `or if the length of the axis is zero and division by zero panics for type A.` on `weighted_mean_axis` because the `MultiInputError::EmptyInput` error happens before any panic. Imo it's better that way.

Can you please check the two first `assert_eq!` of `weighted_sum_dimension_zero`? As I wrote, [0, N] arrays don't make sense to me so I'm not sure what's the right output.

PR #59, opened 2019-12-13
clippy and a bugfix

felix-clark:
Fixed most clippy warnings. Two remain but appear to be in the ndarray `s![]` macro itself.

One warning highlighted that the `ne()` implementation for `NotNone` which actually called `eq()` on the inner field. A default implementation of `ne()` is implied by `eq()`, so the buggy implementation could simply be removed.

### DIFF
--- a/src/entropy.rs
+++ b/src/entropy.rs
@@ -133,7 +133,7 @@ where
     where
         A: Float,
     {
-        if self.len() == 0 {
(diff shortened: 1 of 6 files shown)


LukeMathWalker:
Good catch on `eq` and `neq` - thanks for taking the time! 💯

Issue #62, opened 2020-04-06
Binned statistic similar to scipy.stats.binned_statistic_dd


RolfStierle:
Binned statistic like [scipy.stats.binned_statistic_dd](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_dd.html) similar to [ndarray_stats::histogram::Histogram](https://docs.rs/ndarray-stats/0.3.0/ndarray_stats/histogram/struct.Histogram.html) would allow calculation of more statistical features like weighted histograms, means, variances, min, max etc. of each bin. I would like to add something like that and would be grateful for opinions on how that should look like. @LukeMathWalker 
#### All vs. only one statistic
Is it a good idea to calculate all statisics when a value is pushed to be binned or should only one statistic be calculated which has to be selected beforehand? 
`bs = BinnedStatistic(grid)` vs. `bs = BinnedStatistic(grid, variance)`.
#### Type of output array
Histograms solely count the number of observations in each bin. The default value is zero. For other statistics zero is a valid result event with values in that bin. 
The output could be just the numerical value and comparison with the histogram (through an additional function) allows knowing which bins are empty, or would something similar to `Option<T>` be a good output? 
`[..., 0.0, 0.0, 1.2, ...]` vs `[..., Value(0.0), Empty, Value(1.2), ...]`.

PR #65, opened 2020-04-16
Binned statistic as extension to histogram.

RolfStierle:
#62 Binned statistic implemented as an extension to histogram. 
Binned statistic of 

- sum
- mean
- variance
- standard deviation
- min
- max


xd009642:
@LukeMathWalker @jturner314 do either of you have any opinions on this?

jturner314:
At a high level, this type of functionality seems useful, although I'm not sure that it makes sense to unconditionally update all the statistics on every sample. (I would think there would be cases where you might want e.g. the mean and standard deviation but not the other statistics.) That said, this is a fairly large PR, and I don't have the time to review it at the moment.

PR #75, opened 2021-04-06
Support ndarray 0.15

cassiersg:


xd009642:
So I was going to go for #74 as it also upgrades the other dependencies. However, you have updated the MSRV in the CI I don't suppose if I merge #74 you could just make this PR the CI change, just because there's a bunch of other code changes in that PR for deprecations in libraries that have changed etc

xd009642:
Or you could add the remote from that PR and merge it into this branch, but if you don't know how to do that with git don't worry we can go for the simpler approach :sweat_smile:

xd009642:
Option C:  I can just merge this one and get the other PR to minimise their work to just the dep updates

cassiersg:
@xd009642 Sorry I missed #74 (I'm sending similar PRs to many projects to upgrade to ndarray 15.0 and I forgot to check if there wasn't a PR open.) I'll open a new PR with just the MSRV change on top of #74.

PR #83, opened 2021-08-03
Define and implement a cov_to_corr() method in the CorrelationExt trait

### LINKED ISSUE #82: Implement a `cov_to_corr()` method
I just had a situation where I wanted to convert from a covariance matrix to a correlation matrix. It would be neat if we could do that with one function call.

This is as much a note to myself to implement this (when I get time) as it is a feature request.

multimeric:
Closes #82.


multimeric:
@jturner314

PR #84, opened 2021-08-25
Add argsort methods

jturner314:


pdcook:
Just wanted to bump this PR as this would be an immensely useful addition. I see that the last mention of this was nearly 2 years ago. What more needs to be done for this PR to be merged and is there any way I could help?

bluss:
These could be included in upstream ndarray

Issue #64, opened 2020-04-14
On sorting


lebensterben:
- Currently the `Sort1dExt` trait provides sorting/partitioning methods based on
  `quickselect`. What's the reason for choosing this over `pdqsort` as used in
  `sort_unstable` in STD.
- `histogram/bins.rs` has several calls on `sort_unstable`, and how about a
   feature flag `rayon` for its parallel sort?

Issue #86, opened 2022-03-04
quantile_mut: fatal runtime error: stack overflow


sjackman:
**Description**
`quantile_mut` can fail with the error message:
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
```

**Version Information**
- `ndarray`: 0.15.4
- `ndarray-stats`: 0.5.0
- Rust: 1.58.1

**To Reproduce**
```rust
use ndarray::Array1;
use ndarray_stats::{interpolate::Linear, Quantile1dExt};
use noisy_float::types::{n64, N64};

fn main() {
    {
        let mut array: Array1<N64> = Array1::ones(15300);
        println!("One {}", array.quantile_mut(n64(0.5), &Linear).unwrap());
    }

    {
        let mut array: Array1<N64> = Array1::ones(15600);
        println!("Two {}", array.quantile_mut(n64(0.5), &Linear).unwrap());
    }
}
```

**Observed behavior**
```console
$ cargo run --profile=dev
One 1

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
$ cargo run --profile=release
One 1
Two 1

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
```

**Expected behavior**
```
One 1
Two 1
Three 1
```

**Additional context**
- I'm able to reproduce this issue on both Linux and macOS with the default stack limit of 8 MiB. (`ulimit -s` reports 8192)
- The result is non-deterministic. Re-running the executable can succeed sometimes and fail sometimes. The larger the vector the more likely it is to fail.
- The result depends on whether optimization is enabled.

sjackman:
Exceeding the stack limit suggests to me either infinite recursion or a large data structure is being placed on the heap. I don't see an obvious source of either of those causes in the code. It's possible that more generalized memory corruption could cause this behaviour.
https://github.com/rust-ndarray/ndarray-stats/blob/9ed937302682b1883128d920e0157615b830778b/src/quantile/mod.rs#L437-L498

sjackman:
https://github.com/rust-ndarray/ndarray-stats/blob/b6628c6a1c143532673213c56d46da5fda80cbe8/src/sort.rs#L293
Looks like this function call `_get_many_from_sorted_mut_unchecked` is recursing probably not infinitely but once per size of the vector, perhaps because all the elements in the array are equal. (thanks to @evolvedmicrobe for discovering)

jturner314:
[...]

Issue #72, opened 2021-01-02
Change get_from_sorted_mut signature to be analogous to select_nth_unstable


jturner314:
Rust 1.49 adds a [`select_nth_unstable`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.select_nth_unstable) method to slices which is very similar to [`get_from_sorted_mut`](https://docs.rs/ndarray-stats/0.4.0/ndarray_stats/trait.Sort1dExt.html#tymethod.get_from_sorted_mut), except that it returns a mutable reference to the element and additionally returns slices for the portions before/after the element. It would probably be worthwhile to change the API of `get_from_sorted_mut` to match (and maybe rename it):

```rust
pub fn select_nth_unstable(&mut self, i: usize) -> (ArrayViewMut1<'_, A>, &mut A, ArrayViewMut1<'_, A>)
where
    A: Ord,
    S: DataMut,
```

PR #92, opened 2023-04-07
Remove `Sort1dExt` in favor of `SliceExt`.

n3vu0r:
Reduces worst-case complexity from *O*(*n*²) to *O*(*n* log(*n*)) by using [`ndarray-slice`](https://docs.rs/ndarray-slice).


n3vu0r:
Resolves #64, #72, and #86.

PR #93, opened 2023-06-22
Depend on itertools >= 0.10

sjackman:
Depend on itertools >= 0.10

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ noisy_float = "0.2.0"
 num-integer = "0.1"
 num-traits = "0.2"
 rand = "0.8.3"
-itertools = { version = "0.10.0", default-features = false }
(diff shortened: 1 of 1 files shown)


sjackman:
@jturner314 Thank you for your work on this useful crate! Could you please give this PR a review when you get a chance?

Cargo.toml
```diff
@@ -21,7 +21,7 @@ noisy_float = "0.2.0"
 num-integer = "0.1"
 num-traits = "0.2"
 rand = "0.8.3"
-itertools = { version = "0.10.0", default-features = false }
+itertools = { version = ">=0.10", default-features = false }
```
bluss, line 24: The right change here is to move it to 1.x, not have an open ended interval (then users will get broken builds when itertools has breaking changes.)

We'll get to it if we get through with the ndarray 0.16 update.

PR #100, opened 2024-08-24
Upgrade to codecov-action@v4 and use token

bluss:
Used token from codecov.io (for rust-ndarray org), configured in repo settings.
This seems to fix the code coverage action.


bluss:
Action fails on larger diff in coverage vs 0.5.1 release, but I think it's due to changes in the coverage tool itself too (?)

